PR #3, opened 2016-09-04
Build with Stack

tfausak:


lspitzner:
My stack is still setting up ghc oO But this looks harmless enough :-)

Merged, thanks!

tfausak:
Thank you! This built on my Windows machine. The tests weren't happy, but I think it had to do with line endings. I don't have Cabal set up, so I can't see what it reports for the test suite.

lspitzner:
Tests are fine for me via stack. I have opened #4 as i suspect some newline-handling is not perfect yet in general. Thanks for pointing that out.

PR #13, opened 2017-03-06
Upgrade to Stackage LTS 8.4

tfausak:
This pull request updates the Stack resolver from `nightly-2016-12-04` to `lts-8.4`. It also drops the `multistate-0.7.1.1` and `yaml-0.8.18.1` extra deps because the LTS snapshot includes appropriate versions of those packages. And it drops the `ghc-exactprint` fork because alanz/ghc-exactprint#44 was merged and released in 0.5.3.0, which is in the LTS snapshot. 

With these changes, I am able to clone the repo and run `stack test` successfully.

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,20 +1,16 @@
+resolver: lts-8.4
+
 extra-deps:
-- monad-memo-0.4.1
-- multistate-0.7.1.1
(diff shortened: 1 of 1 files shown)


lspitzner:
thanks!

PR #35, opened 2017-07-07
Fix haddock generation.

alanz:
Otherwise the haddock process complains about malformed haddock comments


lspitzner:
Thanks!

(on a related note, see https://github.com/haskell/haddock/issues/622, and yes, i should have just fixed it instead of/after ranting there.. meh)

Issue #40, opened 2017-08-05
Consider adding `--inplace` option


fmthoma:
Usage:
```
brittany --inplace **/*.hs
```

And since input files can already be specified as positional arguments, the `-i` parameter could be reassigned as shortcut for `--inplace`...

(My current workaround: `ls **/*.hs | xargs -I{} brittany -i {} -o {}`)

lspitzner:
good point.

thinking aloud a bit:
- With your example, brittany would of course need to accept `[FILE..]` and not just `[FILE]`, so just quickly aliasing `--inplace x` to `--input x --output x` is not sufficient. (A great reason for me to refactor [the monstrosity that is the effective `main` function](https://github.com/lspitzner/brittany/blob/master/src-brittany/Main.hs#L130-L255). *whistles innocently*)
- We probably want abort on a per-file basis (syntax error in one file -> all other are still processed).
- Might want error/warning output to be prefixed with the current file, or just print "processing $FILE" for each one in order.
- Might want parallelization.

lspitzner:
PR #59 adds an `--inplace` flag, but due to limitations of the `butcher` library the `brittany --inplace **/*.hs` usecase does not work yet. I'll fix that in `butcher` next.

lspitzner:
I have uploaded `butcher-1.1.1.0` which has `addStringParams`.

@d-dorazio you wanna do the final bit to fix this?

d-dorazio:
@lspitzner sure, but I also have a proposal: specifying input files both with `-i` and as parameters seem a bit verbose to me so I'd remove both `-i` and `-o` and provide an interface similar to what other formatters provide like [`rustfmt`](https://github.com/rust-lang-nursery/rustfmt#running). For example we could add `--write-mode inplace|display` and the files would be specified only as parameters.

lspitzner:
@d-dorazio I agree, the current handling of `--input` vs param(s) can be improved, and `write-mode` makes sense. I don't see the verbosity argument - it only affects the (commandline) docs, but I have no strong opinion on this.

For clarification:

no | --inputs | --outputs | params | behaviour
|-|-|-|-|-
0 | \- | \- | \- | stdin to stdout
1 | \- | \- | params | inplace or according to `--write-mode`
2 | ifile | \- | \- | ifile to stdout
3 | \- | ofile | \- | stdin to ofile
4 | ifile | ofile | \- | ifile to ofile
[...]

PR #59, opened 2017-10-01
add --inplace flag

d-dorazio:
This PR adds support for:
- passing multiple inputs/outputs via the `-i`/`-o` flags;
- the `--inplace` flag as a shortcut for overwriting the input files with the formatted output.

I'm still trying to learn Haskell, so suggestions are appreciated 😄  .

### DIFF
--- a/src-brittany/Main.hs
+++ b/src-brittany/Main.hs
@@ -14,6 +14,7 @@ import qualified Data.Map as Map
 
 import qualified Data.Text.Lazy.Builder as Text.Builder
 
+import           Control.Monad (foldM)
 import           Data.CZipWith
 
 import qualified Debug.Trace as Trace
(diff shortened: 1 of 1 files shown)


lspitzner:
Heh, i was just working on brittany too. I want to take a break now, but will have a look at this later.

src-brittany/Main.hs
```diff
@@ -128,29 +130,36 @@ mainCmdParser helpDesc = do
...
+      Nothing    -> pure ()
+ where
+  run acc io = do
+    res <- io
+    case res of
+      Left _   -> return (Just 1)
+      Right () -> return acc
+
+  nonEmptyList def [] = [def]
+  nonEmptyList _   x  = x
```
lspitzner: In my opinion the case-of that was above before is nicer. This helper obscures the logic by forcing me to jump half a screen down and back up, with little benefit.

There is another source of redundancy I'd rather focus on: The code above effectively checks `inplace` twice and `null outputPaths` twice as well. Merging those two makes more sense imo.

lspitzner:
Thanks! It is merged into `dev` branch.

I'll add a note to #40.

d-dorazio:
@lspitzner the reason why I don't exit with `errNo` is that if more files fail then the exit code would depend on the order of the inputs and imo that'd be strange. I can do it if you feel strongly about it.

I should have addressed all the other comments though.

lspitzner:
I wouldn't have minded depending on the order, but I can understand that you don't like it. In the end you can just not communicate very much in one `Int`. I have merged as-is.

lspitzner:
Btw by using sequence for the `IO` part, making things parallel should be a one-liner.

PR #60, opened 2017-10-04
rework the cli interface

d-dorazio:
I removed the `-i` and `-o` flags so that the inputs are represented by the params only and replaced `--inplace` with `--write-mode [display|inplace]`.

### DIFF
--- a/src-brittany/Main.hs
+++ b/src-brittany/Main.hs
@@ -12,8 +12,12 @@ import qualified Language.Haskell.GHC.ExactPrint.Types as ExactPrint.Types
 import qualified Language.Haskell.GHC.ExactPrint.Parsers as ExactPrint.Parsers
 import qualified Data.Map as Map
 
+import           Text.Read (Read(..))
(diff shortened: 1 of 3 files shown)


src-brittany/Main.hs
```diff
@@ -103,19 +118,24 @@ mainCmdParser helpDesc = do
...
+    ["write-mode"]
+    ""
+    Flag
+      { _flag_help    = Just (PP.text "output mode: [display|inplace]")
+      , _flag_default = Just Display
+      }
   reorderStop
-  inputParam <- addStringParamOpt "PATH" (paramHelpStr "path to input haskell source file")
-  desc       <- peekCmdDesc
+  inputParams <- addParamNoFlagStrings "PATH" (paramHelpStr "paths to input haskell source files")
```
lspitzner: With `addParamNoFlagStrings`, it should be possible to include the params in the reordered block.

lspitzner:
Thanks! I'll see about fixing the help output.

Issue #56, opened 2017-09-21
Turn off all context sensitive alignment in Brittany.


eborden:
Hey @lspitzner, I think `brittany` is a fantastic project, but I'm having trouble getting it to behave the way I'd like. I was able to disable most context sensitive alignment, but I can't seem to rid the world of it. My config is:

```
conf_debug:
  dconf_dump_bridoc_simpl_par: false
  dconf_dump_ast_unknown: false
  dconf_dump_bridoc_simpl_floating: false
  dconf_dump_config: false
  dconf_dump_bridoc_raw: false
  dconf_dump_bridoc_final: false
  dconf_dump_bridoc_simpl_alt: false
  dconf_dump_bridoc_simpl_indent: false
  dconf_dump_annotations: false
  dconf_dump_bridoc_simpl_columns: false
  dconf_dump_ast_full: false
conf_forward:
  options_ghc: []
conf_errorHandling:
  econf_ExactPrintFallback: ExactPrintFallbackModeInline
  econf_Werror: false
  econf_omit_output_valid_check: false
  econf_CPPMode: CPPModeAbort
  econf_produceOutputOnErrors: false
conf_layout:
  lconfig_altChooser:
    tag: AltChooserBoundedSearch
    contents: 3
  lconfig_importColumn: 60
  lconfig_alignmentLimit: 1
  lconfig_indentListSpecial: true
  lconfig_indentAmount: 2
  lconfig_alignmentBreakOnMultiline: false
  lconfig_cols: 120
  lconfig_indentPolicy: IndentPolicyLeft
  lconfig_indentWhereSpecial: true
  lconfig_columnAlignMode:
    tag: ColumnAlignModeDisabled
    contents: []
```

But I still get code like this:

```
+    socialStudiesReadingAssignmentSessions <- getEntities socialStudiesProducts
+                                                          SocialStudiesProductReading
+                                                          getSocialStudiesReadingAssignmentSessions
+    socialStudiesWritingAssignmentSessions <- getEntities socialStudiesProducts
+                                                          SocialStudiesProductWriting
+                                                          getSocialStudiesWritingAssignmentSessions
```

Is there any way to turn this off completely?

lspitzner:
No, this is not implemented (yet). I see you set `lconfig_indentPolicy = IndentPolicyLeft` already, and that setting is planned to do exactly what you want, but I fear the field is currently ignored entirely.

[...]

PR #66, opened 2017-11-19
Support IndentPolicyLeft

eborden:
This PR introduces support for `IndentPolicyLeft` in `brittany`.

This work has proceeded with the intent to couple `IndentPolicyLeft` with `ColumnAlignModeDisabled` to allow a context free formatting style for `brittany`. Decisions on syntactic form follow that philosophy. Tests have been added for this use case. They currently represent a fair amount of duplication, but that should be able to be reduced once inline format annotations are available.

### DIFF
--- a/src/Language/Haskell/Brittany/Internal/Layouters/Expr.hs
+++ b/src/Language/Haskell/Brittany/Internal/Layouters/Expr.hs
@@ -541,7 +541,7 @@ layoutExpr lexpr@(L _ expr) = do
       -- comments before the first let item are moved horizontally with it.
       docSetIndentLevel $ case mBindDocs of
         Just [bindDoc] -> docAltFilter
-          [ ( indentPolicy /= IndentPolicyLeft
+          [ ( True
             , docSeq
               [ appSep $ docLit $ Text.pack "let"
               , appSep $ docForceSingleline $ return bindDoc
--- a/src/Language/Haskell/Brittany/Internal/Layouters/Expr.hs
+++ b/src/Language/Haskell/Brittany/Internal/Layouters/Expr.hs
@@ -853,81 +853,87 @@ layoutExpr lexpr@(L _ expr) = do
           return $ case ambName of
             Unambiguous n _ -> (lfield, lrdrNameToText n, rFExpDoc)
             Ambiguous   n _ -> (lfield, lrdrNameToText n, rFExpDoc)
-      docAlt
+      docAltFilter
         -- singleline
-        [ docSeq
-          [ docNodeAnnKW lexpr Nothing $ appSep $ docForceSingleline rExprDoc
-          , appSep $ docLit $ Text.pack "{"
-          , appSep $ docSeq $ List.intersperse docCommaSep
-                  $ rFs <&> \case
-                    (lfield, fieldStr, Just fieldDoc) ->
-                      docWrapNode lfield $ docSeq
-                            [ appSep $ docLit fieldStr
-                            , appSep $ docLit $ Text.pack "="
-                            , docForceSingleline fieldDoc
-                            ]
-                    (lfield, fieldStr, Nothing) ->
-                      docWrapNode lfield $ docLit fieldStr
-          , docLit $ Text.pack "}"
-          ]
+        [ ( True
+          , docSeq
+            [ docNodeAnnKW lexpr Nothing $ appSep $ docForceSingleline rExprDoc
+            , appSep $ docLit $ Text.pack "{"
+            , appSep $ docSeq $ List.intersperse docCommaSep
+                    $ rFs <&> \case
+                      (lfield, fieldStr, Just fieldDoc) ->
+                        docWrapNode lfield $ docSeq
+                              [ appSep $ docLit fieldStr
+                              , appSep $ docLit $ Text.pack "="
+                              , docForceSingleline fieldDoc
+                              ]
+                      (lfield, fieldStr, Nothing) ->
+                        docWrapNode lfield $ docLit fieldStr
+            , docLit $ Text.pack "}"
+            ]
+          )
         -- wild-indentation block
-        , docSeq
-          [ docNodeAnnKW lexpr Nothing $ appSep rExprDoc
-          , docSetBaseY $ docLines $ let
-              line1 = docCols ColRecUpdate
-                [ appSep $ docLit $ Text.pack "{"
-                , docWrapNodePrior rF1f $ appSep $ docLit $ rF1n
-                , case rF1e of
-                    Just x -> docWrapNodeRest rF1f $ docSeq
-                                     [ appSep $ docLit $ Text.pack "="
-                                     , docForceSingleline $ x
-                                     ]
-                    Nothing -> docEmpty
-                ]
-              lineR = rFr <&> \(lfield, fText, fDoc) -> docWrapNode lfield $ docCols ColRecUpdate
-                [ docCommaSep
-                , appSep $ docLit $ fText
-                , case fDoc of
-                    Just x ->  docSeq [ appSep $ docLit $ Text.pack "="
-                                      , docForceSingleline x
+        , ( indentPolicy /= IndentPolicyLeft
+          , docSeq
+            [ docNodeAnnKW lexpr Nothing $ appSep rExprDoc
+            , docSetBaseY $ docLines $ let
+                line1 = docCols ColRecUpdate
+                  [ appSep $ docLit $ Text.pack "{"
+                  , docWrapNodePrior rF1f $ appSep $ docLit $ rF1n
+                  , case rF1e of
+                      Just x -> docWrapNodeRest rF1f $ docSeq
+                                      [ appSep $ docLit $ Text.pack "="
+                                      , docForceSingleline $ x
                                       ]
-                    Nothing -> docEmpty
-                ]
-              lineN = docSeq
-                [ docNodeAnnKW lexpr (Just AnnOpenC) docEmpty
-                , docLit $ Text.pack "}"
-                ]
-              in [line1] ++ lineR ++ [lineN]
-          ]
+                      Nothing -> docEmpty
+                  ]
+                lineR = rFr <&> \(lfield, fText, fDoc) -> docWrapNode lfield $ docCols ColRecUpdate
+                  [ docCommaSep
+                  , appSep $ docLit $ fText
+                  , case fDoc of
+                      Just x ->  docSeq [ appSep $ docLit $ Text.pack "="
+                                        , docForceSingleline x
+                                        ]
+                      Nothing -> docEmpty
+                  ]
+                lineN = docSeq
+                  [ docNodeAnnKW lexpr (Just AnnOpenC) docEmpty
+                  , docLit $ Text.pack "}"
+                  ]
+                in [line1] ++ lineR ++ [lineN]
+            ]
+          )
         -- strict indentation block
-        , docSetParSpacing
-        $ docAddBaseY BrIndentRegular
-        $ docPar
-            (docNodeAnnKW lexpr Nothing $ rExprDoc)
-            (docNonBottomSpacing $ docLines $ let
-              line1 = docCols ColRecUpdate
-                [ appSep $ docLit $ Text.pack "{"
-                , docWrapNodePrior rF1f $ appSep $ docLit $ rF1n
-                , docWrapNodeRest rF1f $ case rF1e of
-                    Just x -> docSeq [ appSep $ docLit $ Text.pack "="
-                                     , docAddBaseY BrIndentRegular $ x
-                                     ]
-                    Nothing -> docEmpty
-                ]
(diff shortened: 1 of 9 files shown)


eborden:
Actually there was less left in this batch of work then I realised.

This is ready for review.

For posterity this is addressing issues raised in: https://github.com/lspitzner/brittany/issues/56

lspitzner:
In general this looks good. I still want to look at the different layouter changes in detail, and I won't do that today (even when the changes are relatively transparent).

But some first-glance stuff:
- The test setup is reasonable, but I think in conflict with current `dev` branch. Shouldn't be too hard to fix though. Although the proper route is to add a commandline flag for indentpolicy and use the stuff from the `inlineconfig` branch.. but that can be improved at a later point too.
- There is some code-duplication in Expr.hs. Well, admittedly there probably also was before this PR. I'll point out specifics in a closer review.
- Some of the pending testcases that you changed are imo more about horizontal alignment than about indentation. Of course disabling horizontal alignment is in the spirit of #56, too. Maybe needs a bit clarification regarding the purpose of the tests (or maybe it is fine, too. just something i noticed on quick glance.)

tfausak:
I just want to say that I would drop `hindent` for `brittany` in a heartbeat if `brittany` never did any context-sensitive alignment. I feel like `ghc-exactprint` is a more solid foundation than `haskell-src-exts`, but at the same time I don't want code formatted as shown in #56. I can't speak to implementation details, but @eborden's test suite looks exactly like how I would format things by hand.

ElvishJerricco:
Agreed. I have never found myself preferring context sensitive alignment.

src/Language/Haskell/Brittany/Internal/Layouters/Expr.hs
```diff
@@ -554,28 +554,65 @@ layoutExpr lexpr@(L _ expr) = do
...
+              , docLines
+                [ docAddBaseY BrIndentRegular
+                $ docPar
+                  (docLit $ Text.pack "let")
+                  (docSetBaseAndIndent $ docLines $ return <$> bindDocs)
+                , docAddBaseY BrIndentRegular
+                $ docPar
+                  (docLit $ Text.pack "in")
+                  (docSetBaseY $ expDoc1)
+                ]
```
lspitzner: Please remove the duplication here. In the `IndentPolicyLeft` case, the docAlt can move further in, also removing some duplication.

src/Language/Haskell/Brittany/Internal/Layouters/Expr.hs
```diff
@@ -554,28 +554,65 @@ layoutExpr lexpr@(L _ expr) = do
               (docSetBaseY $ expDoc1)
             ]
           ]
-        Just bindDocs@(_:_) -> docAlt
-          [ docLines
-            [ docSeq
-              [ appSep $ docLit $ Text.pack "let"
-              , docSetBaseAndIndent $ docLines $ return <$> bindDocs
+        Just bindDocs@(_:_) -> docAlt $
```
lspitzner: Doesn't the only-one-let-binder-case (above) also need adjustment?

src/Language/Haskell/Brittany/Internal/Layouters/Stmt.hs
```diff
@@ -26,51 +27,58 @@ import {-# SOURCE #-} Language.Haskell.Brittany.Internal.Layouters.Expr
...
+        let letSeq = docSeq
+              [ appSep $ docLit $ Text.pack "let"
+              , docSetBaseAndIndent $ docLines $ return <$> bindDocs
+              ]
+            letRegular = docAddBaseY BrIndentRegular $ docPar
+              (docLit $ Text.pack "let")
+              (docSetBaseAndIndent $ docLines $ return <$> bindDocs)
+        in case indentPolicy of
+          IndentPolicyLeft -> docAlt [letRegular]
+          _ -> docAlt [letSeq, letRegular]
```
lspitzner: Could use `docAltFilter` here too, just to .. stay consistent :D

src/Language/Haskell/Brittany/Internal/Layouters/Stmt.hs
```diff
@@ -26,51 +27,58 @@ import {-# SOURCE #-} Language.Haskell.Brittany.Internal.Layouters.Expr
...
+      Nothing        -> docLit $ Text.pack "let" -- i just tested
+                                -- it, and it is
+                                -- indeed allowed.
+                                -- heh.
+      Just []        -> docLit $ Text.pack "let" -- this probably never happens
+      Just [bindDoc] -> docAlt
+        [ docCols
+          ColDoLet
+          [ appSep $ docLit $ Text.pack "let"
+          , docSetBaseAndIndent $ return bindDoc
```
lspitzner: `docSetBaseAndIndent` there most likely produces ~~hanging indentation~~ (or at least) more-than-usual indentation. Probably need to disable this alt-branch, too.

lspitzner:
A bit of nitpicking: "context sensitive" is ambiguous, but the way i read it here, the only way to be truly context-insensitive would be to remove all `docAlt` nodes (or, equivalently, to disable all but one alternative in every case). Because only then depth in syntax-tree corresponds to indentation exactly, and only then changes to siblings have no chance of affecting the layout of any node. And `IndentPolicyLeft` is far from that, and it is not what we want either, see the [extreme example of truly context-free layouting in my brittany rationale](http://hexagoxel.de/postsforpublish/posts/2017-10-28-brittany-rationale.html#function-and-operator-application).

What we want instead is a compromise between that and the "least newlines possible" layout. Of course the current default may not be the right compromise. Indeed, I am somewhat inclined to making the non-hanging-indentation layout the default. I just don't consider IndentPolicyLeft that different from a theoretical standpoint :p

tfausak:
I don't know if this is the right place to have this discussion, but for me at least "context free" layout means that some expression doesn't depend on the previous expression for its indentation. For example, this is "context sensitive" to me:

``` hs
listToMaybe :: [a]
            -> Maybe a
```

Whereas this is "context free":

``` hs
listToMaybe :: [a]
  -> Maybe a
```

In particular it doesn't mean that every expression is always formatted the same way. Line length is still a concern. But when something forces a newline, the indentation is always "one tab" rather than aligning with the thing above it. 

As I said before, the examples/tests that @eborden added here are exactly how I want things formatted. If there's a term other than "context free" I should be using, I'll be happy to use it instead.

lspitzner:
Ah, so only indentation, and only after newlines. I'd call this "allowing hanging indentation". Perhaps I have too much of the implementation in mind: "indentation after newlines" is just one of the choices of layouting. E.g. if the choices are 1) oneliner 2) multiline with hanging 3) multiline with default indentation, then even if you disable 2) the decision between 1 and 3 is "context sensitive" ("would this overflow 80 cols?" can only be answered in context, for pretty much the same definition of "context" you used.)

(All this does not stand in any way against merging this. I don't mind at all how the test file is named etc. The intend of the issue and the PR are clear and the addition is welcome.)

eborden:
I've updated the code per your inline comments. Would you like me to rebase this over the `dev` branch and change the base of this PR?

lspitzner:
@eborden Well I need these branches merged one way or another. But I can see about the rebase myself, I probably have a better idea of how to handle the conflicts.
(The CI failure is nonsense afaict.)

lspitzner:
Pushed the rebased branch to `eborden-indentpolicyleft`. I was hoping to be able to push directly to the PR.. but what i tried did not work.

There are still two bad testcases. Also I think the stack build does not compile. (The travis CI script is still broken and won't report error correctly.)

eborden:
@lspitzner I'll see about merging that branch into this today.

lspitzner:
I should note that current dev branch (and the one i rebased onto) is slighly broken. nothing to do with the merge, but with the removal of `either` package.

eborden:
@lspitzner Do you always want PRs submitted against the `dev` branch? If so that might be a good thing to outline in a contributors section in `README.md`.

lspitzner:
@eborden indeed, and i already added such a note to the readme a few days ago. bit late for this PR, but the rebase was still manageable :)

would you mind if i rebase again, just so we base this PR on a working commit?

eborden:
@lspitzner Do you want me to give you access to my fork so you can rebase this branch specifically?

eborden:
@lspitzner Added you as a collaborator.

lspitzner:
great, lets see what the CI says. With some luck it will properly report things this time :)

lspitzner:
Ok, that looks right this time; the test failures there are exactly what i see locally. And I think both are just mistakes of the testcases themselves, right?

eborden:
Yup, that seems to be the case. I'll pull this and fix those up in a bit.

src/Language/Haskell/Brittany/Internal/Layouters/Stmt.hs
```diff
@@ -55,38 +55,63 @@ layoutStmt lstmt@(L _ stmt) = do
                                 -- heh.
       Just []        -> docLit $ Text.pack "let" -- this probably never happens
       Just [bindDoc] -> docAltFilter
-        [ ( indentPolicy /= IndentPolicyLeft
+        [ -- let bind = expr
+          ( indentPolicy /= IndentPolicyLeft
```
eborden: Is there a simple way to enable this when the expression is actually a single line? That might be too much complexity thought :smile:

lspitzner: Like this? I think this works.. oh I broke the tests.

lspitzner: But they highlight the changes. And looks approximately correct (although the test coverage is not to be trusted entirely.)

src/Language/Haskell/Brittany/Internal/Layouters/Stmt.hs
```diff
@@ -54,22 +54,22 @@ layoutStmt lstmt@(L _ stmt) = do
...
-          )
+          docCols
+          ColDoLet
+          [ appSep $ docLit $ Text.pack "let"
+          , ( if indentPolicy == IndentPolicyLeft
+              then docForceSingleline
+              else docSetBaseAndIndent
+            )
+            $ return bindDoc
+          ]
```
lspitzner: You could even say this is less complexity :)

lspitzner:
@eborden you probably did not get notified when i commented on the diff comments. Any opinion on the last commit I added? Otherwise I'll assume this is ok and simply merge. (dev branch has diverged already again. not by much, but i really want this merged sooner rather than later.)

eborden:
Oh woops. Sorry about that. That seems like a good approach. So will that fail if it can't force something into a single line?

lspitzner:
Yeah. I'd phrase it as "it restricts the possible layouts for the argument-bridoc to those that fit in a single line" but that probably captures the same idea.

eborden:
Awesome. That sounds prefect to me.

lspitzner:
(a sequence of docLit, docSeparator and (docForceSingleLine x) for arbitrary x thereby only has layouts that are single-line as a whole.)

Great, consider it merged then (I'll wait for CI).

Thanks for the contribution!

I'll make a 0.9.0.0 release that will include this in the next days.

eborden:
Sweet, thanks for helping me through it!

tfausak:
FYI I re-formatted a medium-sized project with `brittany` and it went great: https://github.com/tfausak/rattletrap/commit/e88f4be14de8cad31ecf01f0b97309188b557394?diff=split&w=1

Notes when compared to `hindent`:

- `brittany` puts `::` on a new line instead of on the same line as the identifier. `x\n  :: a` versus `x ::\n  a`. I prefer the way that `brittany` does it.
- In general `brittany` uses a lot fewer newlines without making things cramped, particularly after `->`, `=`, and `<-`. 
- `brittany` puts operators at the beginning of the line, which is pretty much always what I want. 
- `brittany` indents after `let` instead of aligning. Yay! 

There were two things I noticed that felt weird. First, `brittany` always indents after `let` even if it would fit on one line. Also, it does the same thing after `in`. For example:

``` hs
-- this is how i sometimes want things to look
let foo = bar
in foo baz
-- but brittany will space it out like this
let
  foo = bar
in
  foo baz
```

Not really a problem. It eats up some vertical space in the name of consistency. The other weird thing is long expression inside parentheses.

``` hs
-- this is how i would normally format it
(foo
  bar
  baz)
-- this is what brittany does
( foo -- space after opening parenthesis
  bar
  baz
) -- closing parenthesis on new line
```

This feels weird but it's not really a problem. It's more consistent with multi-line tuples or lists, but I don't think I've ever seen Haskell code formatted like this before. On the other hand, it always bugged me that `hindent` formatted tuples like this:

``` hs
( foo
, bar
, baz) -- closing parenthesis on new line
```

So maybe the problem is me 😄

eborden:
Hmm that `let` issue can likely be fixed. There is logic in the statement version that could be ported to the expression form. Could you open an issue?

tfausak:
👉 https://github.com/lspitzner/brittany/issues/73

Issue #70, opened 2017-11-28
Failing on top level template haskell splices


eborden:
For some reason `brittany` is failing to print exact for top level
splices. This may be an issue in `brittany` or `ghc-exact-print`, I'm
not sure. I've added failing tests to highlight this issue. This bug
causes `brittany` to produce syntactically invalid Haskell.

```
1) template haskell top level splice
      expected: Right
                {-# LANGUAGE TemplateHaskell #-}
                deriveFromJSON (unPrefix "assignPost") ''AssignmentPost

      but got: Right
                {-# LANGUAGE TemplateHaskell #-}
                deriveFromJSON (unPrefix "assignPost") ''
```

lspitzner:
I have found the cause of this. The filtering of annotations per top-level decl throws away one annotation connected to the `Unqual` identifier that is in `''AssignmentPost`. I am a bit confused that this part of the syntax tree even _has_ an Annotation, because it cannot have an AnnKey.

I'll investigate this a bit more; however there already would be a brute-force fix: Don't filter annotation for externally-printed top-level decls.

eborden:
Oh awesome. I was just trying to repro this in `ghc-exactprint` the other night. :tada:

lspitzner:
Should be fixed. In theory this fix is a bit sloppy because we should calculate some transitive closure in `extractToplevelAnns`. But as rarely as `annCapturedSpan` seems to be used, I'll take that risk.

PR #71, opened 2017-11-28
Add failing test for template haskell splices

eborden:
For some reason `brittany` is failing to print exact for top level
splices. This may be an issue in `brittany` or `ghc-exact-print`, I'm
not sure. I've added failing tests to highlight this issue. This bug
causes `brittany` to produce syntactically invalid Haskell.

```
1) template haskell top level splice
      expected: Right
                {-# LANGUAGE TemplateHaskell #-}
                deriveFromJSON (unPrefix "assignPost") ''AssignmentPost

      but got: Right
                {-# LANGUAGE TemplateHaskell #-}
                deriveFromJSON (unPrefix "assignPost") ''
```

This test exposes https://github.com/lspitzner/brittany/issues/70


eborden:
This should be getting processed via the wildcard of `ppDecl`.

```
  _         -> briDocMToPPM (briDocByExactNoComment d) >>= layoutBriDoc
```

lspitzner:
> This should be getting processed via the wildcard of `ppDecl`.
correct

This might be similar to https://github.com/alanz/ghc-exactprint/issues/52. Maybe you could try the testing code from that issue? Although I am mildly certain already that this is not on brittany's end.
(of course this is another instance of "would be fixed by brittany properly just supporting  TemplateHaskell/QuasiQuotation". probably a good bit of work though.)

lspitzner:
I added a regression test as part of the fix for #70. (This did not merge anyways.)

PR #78, opened 2017-12-11
ArchLinux installation

damienflament:
I created an [AUR package for Brittany](https://aur.archlinux.org/packages/brittany/).

This PR adds the installation instructions to the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,10 @@ require fixing:
     stack install
     ~~~~
 
+- via `aura` on ArchLinux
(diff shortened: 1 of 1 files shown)


lspitzner:
neat! `aura` is just one AUR client though, perhaps rephrase it to "on archlinux via the AUR, e.g. using `aura`:" ? (although archlinux users probably can deduce as much..)

lspitzner:
Thanks!

PR #91, opened 2017-12-23
stack.yaml: update to lts-10.0

tuncer:


tuncer:
Ran into issues with `cabal new-build`, so I tried stack and, since stack's ghc bindists do not work here, I updated `stack.yaml` to `lts-10.0` in order to use the local host GHC (8.2.2).

tuncer:
One of the CI jobs fails due to the now missing extra deps, which are not needed with lts-10.0. I'd say travis-ci.yml should be adjusted.

tfausak:
I think the typical way to do this is to have a separate `stack-lts-8.0.yaml` and use that in CI instead of setting the `--resolver`.

tuncer:
@tfausak, pushed a commit to that effect. But, looking at the rest of the commented out travis-ci.yml, I'm not sure this is what we want, or that we'd like to carry multiple stack.yaml files at all.

lspitzner:
Renamed the ghc-8.0 stack yaml to `stack-8.0.2.yaml`; Merged into `dev` branch
(github does not see that this is merged, probably because of the rebase and because the commits are not in `master` yet? doesn't really matter.)

Using `--resolver` only makes sense if your package always was on stackage, but that is not really an option for brittany. I think multiple stack yamls are the correct approach.

thanks!

Issue #73, opened 2017-12-03
Format let and in on a single line if they fit


tfausak:
https://github.com/lspitzner/brittany/pull/66#issuecomment-348820992

Using the currently unreleased `IndentPolicyLeft`/`ColumnAlignModeDisabled` config, Brittany formats `let ... in ...` like this:

``` hs
_ =
  let
    longIdentifierForShortValue = 1
  in
    longIdentifierForShortValue + longIdentifierForShortValue
```

I expected both `let` and `in` to be on a single line, like this:

``` hs
_ =
  let longIdentifierForShortValue = 1
  in longIdentifierForShortValue + longIdentifierForShortValue
```

lspitzner:
Indeed, the approach from 882a3b1a7a7851b36fb147580ad36eda2d28abc6 can be applied to the relevant layouter in `Expr.hs` too. I am a bit unsure though which combinations should be accepted, i.e. should the following layouts be possible:

~~~~.hs
_ =
  let singlelinebinder = 1
  in
    some
      multiline
      expression
-- and
_ =
  let
    some long stuff = do
      blah
      blub
  in singleline expression
~~~~

this boils down to permitting layouters
"let (single|multiline) in (single|multiline)" vs
"(let singleline in singleline | let multiline in multiline)"

tfausak:
I would prefer "let (single|multiline) in (single|multiline)". I don't particularly care about them matching; I want them to avoid wasting vertical space.

``` hs
a =
  let it = foo
  in it
-- maybe even `a = let it = foo in it` if everything fits
-- not sure if it's like that already
b =
  let
    it = do
      foo
      bar
  in it
c =
  let it = foo
  in -- maybe start long expression here instead of next line
    do
      it
      bar
d =
  let
    it = do
      foo
      bar
  in
    do
      it
      baz

eborden:
@lspitzner I believe you can close this with the merger of #97

PR #97, opened 2017-12-31
Format let and in on a single line if they fit

eborden:
This addresses #73 

The following is wasteful of vertical space:

```
_ =
  let
    longIdentifierForShortValue = 1
  in
    longIdentifierForShortValue + longIdentifierForShortValue
```

We should format it on two lines if possible.

```
_ =
  let longIdentifierForShortValue = 1
  in longIdentifierForShortValue + longIdentifierForShortValue
```

This commit also allows for a mix of variations:

```
_ =
  let
    longIdentifierForShortValue = 1
  in longIdentifierForShortValue + longIdentifierForShortValue

_ =
  let longIdentifierForShortValue = 1
  in
    longIdentifierForShortValue + longIdentifierForShortValue
```

### DIFF
--- a/src/Language/Haskell/Brittany/Internal/Layouters/Expr.hs
+++ b/src/Language/Haskell/Brittany/Internal/Layouters/Expr.hs
@@ -531,6 +531,9 @@ layoutExpr lexpr@(L _ expr) = do
     HsLet binds exp1 -> do
       expDoc1 <- docSharedWrapper layoutExpr exp1
       mBindDocs <- layoutLocalBinds binds
+      let
+        whenIndentLeftOr x y =
+          if indentPolicy == IndentPolicyLeft then x else y
       -- this `docSetIndentLevel` might seem out of place, but is here due to
       -- ghc-exactprint's DP handling of "let" in particular.
       -- Just pushing another indentation level is a straightforward approach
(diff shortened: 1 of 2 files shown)


src/Language/Haskell/Brittany/Internal/Layouters/Expr.hs
```diff
@@ -531,46 +531,46 @@ layoutExpr lexpr@(L _ expr) = do
...
+                      $ return bindDoc
+                      ]
+                  , docAddBaseY BrIndentRegular
+                  $ docPar
+                    (appSep $ docLit $ Text.pack "let")
+                    (docSetBaseAndIndent $ return bindDoc)
+                  ]
+              , docAlt
+                  [ docSeq
+                      [ ifIndentLeftElse id appSep $ docLit $ Text.pack "in "
```
lspitzner: uhm, `appSep` should append a separator (space) if required. So `appSep "in "` is redundant, and also there should be no reason not to unconditionally use `appSep` here. (and i realize this was not introduced by this diff - the same redundancy was in the before-state.)

Or was there some issue around a comment following `in`?.. lets just test. could you try if replacing `ifIndentLeftElse id appSep $ docLit $ Text.pack "in "` with `appSep $ docLit $ Text.pack "in"` works?

lspitzner:
looks good (otherwise)!

src/Language/Haskell/Brittany/Internal/Layouters/Expr.hs
chreekat: Shooting from the hip here, but maybe this makes xxx and yyy line up in
```haskell
let xxx = ...
in  yyy ...
```
?

I'd check, but my network connection won't really let me at present.

lspitzner: @chreekat oh, right. I missed that detail. I guess then this is fine.

lspitzner:
thanks!

eborden:
Thanks for finishing this one off @lspitzner

PR #105, opened 2018-01-13
Make record declarations respect alignment

eborden:
This commit introduces tests for aligned record declarations. It
utilizes `docCols` and introduces a new `ColSig` for for record
declarations.


eborden:
Hmm looks like CI failure is a difference between 8.0.x and 8.2.x. Should I pull out some CPP or would you rather drop support for 8.0.x?

eborden:
Oh also this is part of #47

lspitzner:
thanks!

Issue #101, opened 2018-01-04
Syntactically invalid result on Esqueleto code


MaxGabriel:
Running `brittany /Users/maximiliantagher/Documents/Mercury/hs/mercury-web-backend/src/Handler/Home.hs` gives me: "ERROR: brittany pretty printer returned syntactically invalid result.", even though the code compiles fine.

Source file:

```haskell
module Handler.Home where

import Import
import qualified Database.Esqueleto      as E
import           Database.Esqueleto      ((^.))
import Data.String.Conversions

getHomeR :: Handler Value
getHomeR = do
    result <- runDB $
        E.select $
        E.from $ \(u `E.InnerJoin` ou `E.InnerJoin` o) -> do
        E.on (ou ^. OrganizationUserOrganizationId E.==. o ^. OrganizationId)
        E.on (u ^. UserId E.==. ou ^. OrganizationUserUserId)
        return (u, ou, o)
    putStrLn $ cs $ show result
    returnJson [1 :: Integer]

```

If you remove the Esqueleto code (the `result <- ...` block of code), brittany runs successfully.

Version:

```
maximiliantagher@Maximilians-MacBook-Pro ~> brittany --version
brittany version 0.9.0.0
Copyright (C) 2016-2017 Lennart Spitzner
There is NO WARRANTY, to the extent permitted by law.
```

OS: macOS High Sierra 10.13.2
GHC: Should be 8.2.2 since brittany was installed with LTS 10.

Installation method: `stack install brittany --resolver lts-10.0`

I also tried with my older version of brittany, which was 0.8.something, and compiled with GHC 8.0.x.

Brittany configuration:

```yaml
conf_debug:
  dconf_roundtrip_exactprint_only: false
  dconf_dump_bridoc_simpl_par: false
  dconf_dump_ast_unknown: false
  dconf_dump_bridoc_simpl_floating: false
  dconf_dump_config: false
  dconf_dump_bridoc_raw: false
  dconf_dump_bridoc_final: false
  dconf_dump_bridoc_simpl_alt: false
  dconf_dump_bridoc_simpl_indent: false
  dconf_dump_annotations: false
  dconf_dump_bridoc_simpl_columns: false
  dconf_dump_ast_full: false
conf_forward:
  options_ghc:
  - -XLambdaCase
  - -XMultiWayIf
  - -XGADTs
  - -XPatternGuards
  - -XViewPatterns
  - -XRecursiveDo
  - -XTupleSections
[...]

PR #107, opened 2018-01-15
Fix infix constructor pattern matching for normal constructors

eborden:
Brittany was previously only supporting symbol based infix constructors. It
is common in some libraries (for example Esqueleto) to pattern match on
normal constructors as infix. Brittany was failing in this case by not
wrapping the constructor name in back ticks/spaces. Backticks and spaces
have been added in the case where the constructor contains any alpha
characters.

This addresses #101


eborden:
This could be moved to the dev branch, but it seems like a simple enough fix to make it in to `master` via a point release.

lspitzner:
Thanks! I think using `lrdrNameToTextAnn` instead of `lrdrNameToText` is a better solution, as it does not depend on `isAlpha`. And the spaces should be added unconditionally, to be consistent with operator space handling in other cases.

Please change to `dev`. I agree with making a minor release for this, but the other stuff currently on `dev` deserves to be included, too :)

eborden:
Hmm, getting some strange business with trying to rebase off `upstream`. Seems some commits have snuck in.

eborden:
Looks like OSX made it through the tests, but then timed out.

eborden:
@lspitzner do you see any blockers on this PR?

lspitzner:
Sorry for the long delay. I have effectively merged this PR now, although I think github won't see this because I rebased/cherry-picked again. But the commits are now on `dev`. I am working on a fix for #116 and will do a release once that is done.

(And you are right, the test failure is irrelevant.)

PR #117, opened 2018-01-24
Add Vim / Neovim plugin

jez:
I saw that you started an editor integration section, and
thought that it might benefit from a Vim / Neovim section \o/

Thanks for this program by the way 🙂


lspitzner:
thanks!

Issue #122, opened 2018-02-24
brittany doesn't format multibyte strings nicely


pythonissam:
```bash
$ brittany --version
brittany version 0.9.0.0
Copyright (C) 2016-2017 Lennart Spitzner
There is NO WARRANTY, to the extent permitted by law.
```

# Example
```haskell
f xs
  | xs == "a" = True
  | xs == "abc" = True
```

will be formatted correctly like below:

```haskell
f xs | xs == "a"   = True
     | xs == "abc" = True
```

Though

```haskell
f xs
  | xs == "あ" = True
  | xs == "abc" = True
```

becomes

```haskell
f xs | xs == "あ"   = True
     | xs == "abc" = True
```

I guess brittany doesn't handle multibyte strings.

lspitzner:
I see, that is a bit unfortunate. However it does not seem to be connected to "multibyte" aspect, but rather to the fact that even with monospace fonts, the glyph for "あ" is larger than other glyphs. The alignment functionality in brittany really only makes sense when all glyphs have the same size.

So the only fix I can think of is to disable alignment: Either globally via
`_lconfig_columnAlignMode = ColumnAlignModeDisabled` or perhaps we could implement something that disables only in relevant cases? I don't really know and somewhat doubt that such cases could be detected reliably though, because it is more the font's property than the codepoint's.

tfausak:
How _should_ this be aligned?

``` hs
f x = case x of
  "1" -> () -- U+0031
  "２" -> () -- U+FF12
  _   -> ()
```

All of the arrows are aligned. They start at column 8. But the characters themselves aren't the same visual width.

PR #123, opened 2018-02-26
Fix issue #122

waddlaw:
Ref: #122 

My idea is replace `length` to `elasticLength` that it is calclate length is two if `**non**-ascii characters`.

Implement `elasticLength` like this:

```haskell
-- |
-- >>> elasticLength "あ"
-- 2
-- >>> elasticLength "abc"
-- 3
-- >>> elasticLength "aあa"
-- 4
elasticLength :: Text -> Int
elasticLength = Text.foldl' (\len c -> if isAscii c then len + 1 else len + 2) 0
```

I will not break anything so far.
@lspitzner Could you merge this?

### DIFF
--- a/brittany.cabal
+++ b/brittany.cabal
@@ -108,6 +108,7 @@ library {
     , cmdargs >=0.10.14 && <0.11
     , czipwith >=1.0.0.0 && <1.1
     , ghc-boot-th >=8.0.1 && <8.3
+    , text-latin1 >= 0.3 && < 0.4
     }
   default-extensions: {
     CPP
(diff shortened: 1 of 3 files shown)


pythonissam:
@lspitzner This PR seems to have fixed the issue. Would you consider merging it?

lspitzner:
Thanks for this contribution!

1. What is the additional dependency for?
2. This does not work for me. With my font, the those glyphs do not have a width of 2 ascii glyphs, so this does not give proper alignment still. Just pointing this out; If it works for you and other users I can merge still.
3. Please test this with some other alphabets/codepages and report what you tested (I don't want an additional cabal testsuite or anything, but some example that I can test manually). I highly suspect that stuff will break, because isAscii is a rather poor estimation of glyph size in general. And any unicode stuff should be tested with more than just two languages.

lspitzner:
Closing due to inactivity, and because afaict this change makes things worse for many other languages (e.g. consider greek or russian).

waddlaw:
@lspitzner I understand. Thank you!

PR #132, opened 2018-03-29
Fix some hlint suggestions

sergv:


lspitzner:
How configurable is `hlint`?

There are some good changes in there, but many of these changes that save a few characters feel rather code-golfy and not worth the effort in my opinion. Take the removal of redundant `do`s: For many of those cases, if the relevant code gets changed in the future, it is very likely that the `do` will become non-redundant. And you gain nothing from saving a whitespace and two characters.

I approve of anything that has to do with alignment, as that has a significant effect on readability.

For other stuff I really don't see the point in even reviewing the individual cases. Sorry.

lspitzner:
I'd like to also give an example of something that would be rather useful to improve in regards to general readability of the layouter's code. This function:

~~~~.hs
docAltFilter :: [(Bool, ToBriDocM BriDocNumbered)] -> ToBriDocM BriDocNumbered
docAltFilter = docAlt . map snd . filter fst
~~~~

is used rather often, but the argument being a list of tuples produces rather ugly code/formatting wherever it is used. There is a different approach that was also mentioned in one `reddit/r/haskell` discussion recently. What if instead of

~~~~.hs
myDoc = docAltFilter
  [ (,) someCondition
  $ some
    large
    expression
  , (,) otherwise
  $ other
    large
    expression
  ]
~~~~

which is rather annoying to read, we could write something like

~~~~.hs
myDoc = runFilteredAlternative $ do
  addAlternativeCond someCond $ some
    large
    expression
  addAlternative $ other
    large
    expression
~~~~

This seems to much clearer syntactically, with the downside of having a bit more mental overhead (because we use yet another Monad). But I think that trade-off is worth it, especially with speaking names `runFilteredAlternative` and `addAlternativeCond` (and not just `runWriter` - it is mostly a Writer, right?).

Maybe you could explore this a bit?

sergv:
Thanks for prompt repsonse! The `hlint` is pretty configurable - each suggestion can be disabled individually and config can be added into the repository.

Regarding `do`s I'd agree as they're pretty harmless. Yet, on the other hand, redundant parens, `$`s or replacement of `fmap f $` by `f <$>` does feel like a good thing - the less operators the better. I do understand that there's quite a few changes and it's tiresome to review them one by one. But on the other hand some changes help to bring constistency to the code, which IMO does improve readability further.

Regarding `docAltFilter` I did notice this pattern and it strikes me as unwieldy. Let me try and see whether I can come up with a nice monad like you're proposing. It indeed does sound like a simple `Writer` monad at its core.

sergv:
@lspitzner Returning to your example for replacement of `docAltFilter`. What do you think about using operators to tie condidion and expression? E.g. instead of 
```hs
myDoc = runFilteredAlternative $ do
  addAlternativeCond someCond $ some
    large
    expression
  addAlternative $ other
    large
    expression
```
write something like
```hs
myDoc = runFilteredAlternative $ do
  someCond  |-> some
    large
    expression
  otherwise |-> other
    large
    expression
```

lspitzner:
> Thanks for prompt repsonse! The hlint is pretty configurable - each suggestion can be disabled individually and config can be added into the repository.

Could you disable the redundant-do part then? I think all other hlint changes I am fine with.

> write something like
>
> ~~~~.hs
> myDoc = runFilteredAlternative $ do
>  someCond  |-> some
>    large
>    expression
>  otherwise |-> other
>    large
>    expression
> ~~~~

It looks cool for sure, however: This particular operator looks a lot like case/guard syntax, where only ever one of the alternatives gets selected - but that is not the case here. And having to use `otherwise` feels like a downside too, when we really mean to express "always". I am not entirely convinced yet :-)

sergv:
I have split initial commit into single commit for each hint, added hlint config that disables `Redundant do` and a couple of other suggestions and replaced `docAltFilter` with `runFilteredAlternative` as suggested.

src/Language/Haskell/Brittany/Internal/LayouterBasics.hs
```diff
@@ -415,8 +423,24 @@ docExt x anns shouldAddComment = allocateNode $ BDFExternal
 docAlt :: [ToBriDocM BriDocNumbered] -> ToBriDocM BriDocNumbered
 docAlt l = allocateNode . BDFAlt =<< sequence l
 
-docAltFilter :: [(Bool, ToBriDocM BriDocNumbered)] -> ToBriDocM BriDocNumbered
-docAltFilter = docAlt . map snd . filter fst
+newtype CollectAltM a = CollectAltM (Writer.Writer [ToBriDocM BriDocNumbered] a)
+  deriving (Functor, Applicative, Monad)
+
+addAlternativeCondM :: Bool -> CollectAltM (ToBriDocM BriDocNumbered) -> CollectAltM ()
```
lspitzner: I don't really see the usecase for this, and one might expect the same pattern as with `when/whenM`:

~~~~.hs
when :: Applicative f => Bool -> f () -> f ()
whenM :: Monad m => m Bool -> m () -> m ()
~~~~

(Also it is not used anywhere afaict.)

src/Language/Haskell/Brittany/Internal/Layouters/Decl.hs
```diff
@@ -313,253 +313,231 @@ layoutPatternBindFinal alignmentToken binderDoc mPatDoc clauseDocs mWhereDocs ha
...
-       ]
-    ++ -- two-line solution + where in next line(s)
-       [ ( True 
-         , docLines
-         $  [ docForceSingleline
-            $ docSeq (patPartInline ++ [guardPart, return binderDoc])
-            , docEnsureIndent BrIndentRegular $ docForceSingleline $ return body
+  runFilteredAlternative $ do
+
+    let wherePart = case mWhereDocs of
```
lspitzner: could you move this let up in group with the previous let-binds?

src/Language/Haskell/Brittany/Internal/Layouters/IE.hs
```diff
@@ -122,24 +119,20 @@ layoutLLIEs enableSingleline llies = do
...
-      [ (not hasComments, docLit $ Text.pack "()")
-      , ( hasComments
-        , docPar (docSeq [docParenLSep, docWrapNodeRest llies docEmpty])
-                 docParenR
-        )
-      ]
-    (ieDsH:ieDsT) -> docAltFilter
-      [ ( not hasComments && enableSingleline
-        , docSeq
+    [] -> runFilteredAlternative $ do
```
lspitzner: Can factor out `runFilteredAlternative $` out of the `case`s.

src/Language/Haskell/Brittany/Internal/Layouters/Decl.hs
```diff
@@ -313,253 +313,231 @@ layoutPatternBindFinal alignmentToken binderDoc mPatDoc clauseDocs mWhereDocs ha
...
+          $ docLines
+          $  [ docSeq (patPartInline ++ [guardPart, return binderDoc])
+             , docEnsureIndent BrIndentRegular
+             $ docNonBottomSpacing
+             $ docAddBaseY BrIndentRegular
+             $ return body
+             ]
+          ++ wherePartMultiLine
+
+      _ -> return ()
```
lspitzner: add comment

~~~~.hs
return () -- no alternatives exclusively when `length clauseDocs /= 1`
~~~~

src/Language/Haskell/Brittany/Internal/LayouterBasics.hs
sergv: Indeed it's not used anywhere. I think I tried to use it somewhere but found a way without it.

src/Language/Haskell/Brittany/Internal/Layouters/Decl.hs
sergv: Done.

src/Language/Haskell/Brittany/Internal/Layouters/IE.hs
sergv: Done.

src/Language/Haskell/Brittany/Internal/Layouters/Decl.hs
sergv: Done.

sergv:
I have addressed your suggestions. Please review.

lspitzner:
Thanks! I'd never have noticed that cabal file duplication :-)

PR #160, opened 2018-07-09
Add note about Atom editor integration in README, fixes #62

### LINKED ISSUE #62: Use with the Atom editor - possible?
I'm currently using hindent, which integrates very well with Atom (as does stylish-haskell). Is there a similar story for brittany? Anybody here using it with Atom to prettify on save?

Tehnix:


lspitzner:
Ah, I read your comment but somehow overlooked the PR earlier. Thanks!

PR #168, opened 2018-08-06
Fix a brittany misspelling

robx:
Trivial I know, but I keep typoing brittany myself so this one really jumped out...


lspitzner:
Heh, thanks! :-)

PR #172, opened 2018-08-11
Add note about Emacs integration in README

disya2:
Thanks for the great tool! I've been using it for a while and I think it does an excellent job. Using it from Emacs is straightforward. I made a change to README file to reflect my setting.

A note: when I hit "save" the editor also removed some trailing white spaces. Please let me know if it is a problem and I'll redo the change to exclude those.


ElvishJerricco:
Why not use stuff like `shell-command-on-region`? i.e. send the Emacs buffer through stdin instead of the file through CLI, so that you don't have to save before formatting, and so you can format only a region.

disya2:
`shell-command-on-region` is totally fine when you want to format a region. And I used to have that function as well. 
However, I found that in most cases I just want the whole file to be always formatted. I type in a few lines, hit the keys to verify the result and continue writing. I also have that function hooked on buffer save. And since formatting is idempotent there's no harm to run that multiple times, I think.

ElvishJerricco:
@disya2 Right but there must surely be something like `shell-command-on-region` that applies to the whole file? It just seems wrong to work on the file level when the user is working on the buffer level.

ChickenProp:
Note that haskell-mode provides `haskell-mode-stylish-buffer` and `haskell-mode-buffer-apply-command` which don't require a save. I have

```elisp
(eval-after-load 'haskell-mode
  (lambda ()
    (progn
      (setq haskell-mode-stylish-haskell-path "brittany")
      (define-key haskell-mode-map "\C-c\C-f" #'haskell-mode-stylish-buffer))))
```

lspitzner:
I don't use emacs, so I have no comment on how good of an approach this is. Still, I'd prefer to not drop the proposed code entirely if it works, and nobody has bothered to write something better yet, even if it has flaws.

I propose to put the code snippet in a separate file in the docs folder, together with a note of potential downsides, and add a link to this from the readme, under the emacs header.

@disya2 thanks for proposing this. Even though you closed this, would you be fine with the above?

PR #178, opened 2018-09-11
Add check mode for use by test suites

mightybyte:
This pull request adds a mode makes no changes to files, but returns 0 (success) when no changes would be made and 1 (failure) when changes would be made.  This mode is useful for enforcing the use of brittany on a project without imposing things on the dev workflow.  You can simply run this mode in CI and it will fail if brittany needs to be run.


lspitzner:
Looks good, thanks!

CMCDragonkai:
Can this be added into the documentation, how to use in the command line? I'd like to integrate this into my Gitlab precommit hooks.

Issue #176, opened 2018-08-31
Record attribute formatting doesn't respect IndentPolicyLeft


pbrisbin:
(As tested on https://hexagoxel.de/brittany/)

Given:

```hs
-- brittany-next-binding { lconfig_indentPolicy: IndentPolicyLeft }

record :: Record
record = Record
  { rProperties =
    [ "foo" .= "bar"
    , "foo" .= "bar"
    , "foo" .= "bar"
    , "foo" .= "bar"
    , "foo" .= "bar"
    , "foo" .= "bar"
    , "foo" .= "bar"
    , "foo" .= "bar"
    ]
  }
```

I would expect **no re-formatting** to occur.

Instead, we get this:

```hs
-- brittany-next-binding { lconfig_indentPolicy: IndentPolicyLeft }

record :: Record
record = Record
  { rProperties = [ "foo" .= "bar"
                  , "foo" .= "bar"
                  , "foo" .= "bar"
                  , "foo" .= "bar"
                  , "foo" .= "bar"
                  , "foo" .= "bar"
                  , "foo" .= "bar"
                  , "foo" .= "bar"
                  ]
  }
```

Interestingly, normal bindings seem to have the reverse issue. Given this:

```hs
properties =
  [ "foo" .= "bar"
  , "foo" .= "bar"
  , "foo" .= "bar"
  , "foo" .= "bar"
  , "foo" .= "bar"
  , "foo" .= "bar"
  , "foo" .= "bar"
  , "foo" .= "bar"
  ]
```

And the default `indentPolicy`, I would expect that to become

```hs
properties = [ "foo" .= "bar"
             , "foo" .= "bar"
             , "foo" .= "bar"
             , "foo" .= "bar"
             , "foo" .= "bar"
             , "foo" .= "bar"
             , "foo" .= "bar"
             , "foo" .= "bar"
             ]
```

But it does not, it stays where it is.

So I think there are two bugs here:

1. Bindings vs record attributes do not format consistently
1. `IndentPolicyLeft` is not respected for record attributes

I personally care only about (2) since the (arguably broken) behavior for bindings results in the formatting I prefer -- so that's how I titled the issue.

/cc @eborden

PR #184, opened 2018-09-16
Consolidate record expression layouter

eborden:
Both record construction and update layouting have very similar
constructions. These each had their own layouter with slightly different
variations. Variations here lead to subtly different bugs in layout for
nearly identical syntactic forms.

The record update logic is more advanced and respects `IndentPolicyLeft`.
Instead of keeping these layouters distinct we can consolidate
construction logic into the update logic. This results in a smaller
volume of code and more uniform layouting of syntax for these similar
forms.

Record constructors with fields and wildcards are not included in this
consolidation. A `TODO` has been left to handle this consolidation later.

This fixes https://github.com/lspitzner/brittany/issues/176


lspitzner:
thanks!

PR #185, opened 2018-09-17
Append input path name to UnknownNode errors

### LINKED ISSUE #175: Include filename in parse error messages
When one runs brittany on multiple files and one of them errors, you can't tell which file had the error:

```console
% brittany --write-mode inplace **/*.hs
ERROR: encountered unknown syntactical constructs:
HsSpliceE{}
HsSpliceE{}
ERROR: encountered unknown syntactical constructs:
HsSpliceE{}
```

When the file list is _very_ large, it becomes impossible to track things down.

This is probably because the error message format was decided before `--write-mode inplace` existed, so operating over multiple files didn't make much sense. Now that we have that though, I would like to request that the filename be included in these errors.

5outh:
Appends the name of the input file to any `UnknownNode` errors. This is useful when running `brittany` over a set of source files.

Example output:

```shell
ERROR: encountered unknown syntactical constructs when parsing file "library/Application.hs":
HsSpliceE{}
ERROR: encountered unknown syntactical constructs when parsing file "library/Foundation.hs":
HsSpliceE{}
ERROR: encountered unknown syntactical constructs when parsing file "library/FrontRow/Api/ActivityFeed/Handler.hs":
HsBracket{}
HsBracket{}
HsBracket{}
ERROR: encountered unknown syntactical constructs when parsing file "library/FrontRow/Api/DistrictAssessment/Handler/Answers.hs":
HsBracket{}
```

```shell
ERROR: encountered unknown syntactical constructs when parsing stdin:
HsSpliceE{}
```

Closes #175.

### DIFF
--- a/src-brittany/Main.hs
+++ b/src-brittany/Main.hs
@@ -285,6 +285,8 @@ coreIO putErrorLnIO config suppressOutput inputPathM outputPathM =
             return $ Right True
           CPPModeNowarn -> return $ Right True
         else return $ Right False
+    let
+      inputPathName = maybe "stdin" (("file " <>) . show) inputPathM
     (parseResult, originalContents) <- case inputPathM of
       Nothing -> do
         -- TODO: refactor this hack to not be mixed into parsing logic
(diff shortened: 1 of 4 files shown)


lspitzner:
I was so free to push another commit. Would you mind confirming if this still works as desired for some multiple-file testcase?

(and tell if you prefer the previous behaviour - i just thought "wait, we really should have the full srcspan info available in `ErrorUnknownNode`" and used that.)

5outh:
@lspitzner All works on my end. Thanks!

lspitzner:
thanks!

PR #186, opened 2018-09-17
Unboxed tuples

sergv:
I noticed that indenting snipptes like
```haskell
case foo# of
  (# bar#, baz# #) -> ...
```

will produce output like
```haskell
case foo# of
  (#bar#, baz##) -> ...
```

which is invalid. I decided to put spaces around delimiters in `(#...#)` unconditionally to avoid having to check whether names have hashes. This style also seems more prevalent from my non-representative sample.

I also enabled running literate tests in parallel on my machine and though it could be useful for others. Please let me know what you thing about using all available cores to run the tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -292,7 +292,7 @@ script:
     canew)
       better_wait cabal new-build -j$JOBS --ghc-options="-j1 +RTS -M500M -RTS" --disable-tests --disable-benchmarks
       better_wait cabal new-build -j$JOBS --ghc-options="-j1 +RTS -M500M -RTS" --enable-tests --enable-benchmarks
-      cabal new-test --ghc-options="-j1 +RTS -M500M -RTS"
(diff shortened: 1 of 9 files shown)


ElvishJerricco:
I think doing this unconditionally is pretty common style anyway, so I think this is a good change

lspitzner:
Thanks! I agree that it makes sense to consistently use spaces for unboxed tuples.

I have pushed another commit with minor cleanups. Should be ready to merge.

Regarding the parallel testing, I am a bit confused about the failing test. The failure indicates a performance issue (and might be due to heavy load on travis) but then it is for a testsuite that you did not even touch. Plus it is only new-build that is failing, so I wonder if new-build is just somehow picking up the flag or running entire suits in parallel or something. Maybe I have to force serial test execution via `-j1` for new-build on travis? If all that does not work I might prefer to revert that bit, but let's see first.

PR #198, opened 2018-11-12
Add type fam instance formatting

ruhatch:
This adds proper formatting of type family instances. It also refactors the common pattern found in signatures, type synonyms, and type family instances, into a single function.


eborden:
Will this have support for closed type families?

ruhatch:
@eborden No, this only adds support for the `type instance ...` syntax (as well as the `type` syntax inside type class instances).

lspitzner:
Thanks!

PR #213, opened 2019-01-26
Add build instructions for nix

bsima:
This works on my NixOS machine, and it should work everywhere else too, since i've pinned the nixpkgs version in `pkgs.nix`. The `default.nix` file supports both `nix build` and `nix-shell` automatically. You can also do e.g. `nix-shell --argstr compiler ghc844` to change the compiler; I have it set to `ghc822` as the default because that's what worked for me when building.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ local/
 cabal.sandbox.config
 cabal.project.local
 .ghc.environment.*
+result
--- a/README.md
+++ b/README.md
@@ -94,6 +94,12 @@ log the size of the input, but _not_ the full input/output of requests.)
     aura -A brittany
     ~~~~
 
+- via `nix`:
+    ~~~.sh
+    nix build  # or 'nix-build'
+    nix-env -i ./result
+    ~~~
+
 # Editor Integration
 
 #### Sublime text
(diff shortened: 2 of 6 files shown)


ElvishJerricco:
Note that the use of `developPackage` means that this Nix file cannot be imported by any other project that intends to use `nix-shell`.

bsima:
Will Fancher <notifications@github.com> writes:

> Note that the use of `developPackage` means that this Nix file cannot
> be imported by any other project that intends to use `nix-shell`.

Good point Will. Commit e67a46f2649ee53380f2a7d933588e8587ba0ecd is an
attempt to address this.

ElvishJerricco:
@bsima How does that address this? `release.nix` still uses `callPackage ./shell.nix`, and `shell.nix` uses `developPackage`.

bsima:
Will Fancher <notifications@github.com> writes:

> @bsima How does that address this? `release.nix` still uses
> `callPackage ./shell.nix`, and `shell.nix` uses `developPackage`.

Users can `import ./default.nix` to get a nix expression without `developPackage`

lspitzner:
I am gonna optimistically assume that the complain has indeed been resolved. Have not tested, but if there are issues we can always change things again later.

Thanks for the PR.

Issue #214, opened 2019-02-04
Arguments with comments walk to the right


matt-noonan:
In a multi-line function application with line comments on some arguments, the indent level increases after each commented line. For example:

```haskell
foo = bar
  arg1 -- this is the first argument
  arg2 -- this is the second argument
  arg3 -- this is the third argument, now I'll skip one comment
  arg4
  arg5 -- this is the fifth argument
  arg6 -- this is the sixth argument
```

formats to

```haskell
foo =
  bar arg1 -- this is the first argument
           arg2 -- this is the second argument
                arg3 -- this is the third argument, now I'll skip one comment
                     arg4 arg5 -- this is the fifth argument
                               arg6 -- this is the sixth argument
```

tfausak:
Closing as fixed by #215.

PR #215, opened 2019-02-04
Disable single-line HsApp with argument comments.

matt-noonan:
This resolves issue #214.


lspitzner:
Ah, a nice one-liner. Thanks! Will merge after travis.

PR #216, opened 2019-02-05
Explicitly handle empty HsCase and HsLamCase.

### LINKED ISSUE #203: Braces for EmptyCases
As mentioned over at https://github.com/haskell/haskell-ide-engine/issues/967, it may be preferable to have EmptyCases be explicit, that is:

~~~~.hs
absurd a = case a of {}
-- and the lambdacase-variant
absurd = \case {}
~~~~

I think this is sensible, at least behind a config flag, and perhaps even by default.

matt-noonan:
This resolves #203 by adding `{}` to empty `case` and `\case` expressions.


lspitzner:
Thanks!

PR #244, opened 2019-06-23
Add support for OverloadedLabels

### LINKED ISSUE #243: Support -XOverloadedLabels
The `HsExpr` constructor is `HsOverLabel`, and a valid syntax example is:
```
foo = #bar 
```

eborden:
Closes #243 

`OverloadedLabels` is a simple enough extension to parse and format. It
is becoming more common with use of `generic-lens`. Since it can be
treated as a `HsVar` its implementation only requires using `docLit`,
along with some marshalling for dealing with `FastString`.

### DIFF
--- a/src-literatetests/14-extensions.blt
+++ b/src-literatetests/14-extensions.blt
@@ -130,3 +130,13 @@ func = do
   hello
   |]
   pure True
+
+###############################################################################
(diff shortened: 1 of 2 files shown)


lspitzner:
Thanks!

PR #279, opened 2020-02-04
Bumps neat-interpolation upper bound

jkachmar:
Addresses commercialhaskell/stackage#5119.

I've tested this locally and verified that `brittany` compiles and tests pass with `neat-interpolation-0.4`.


lspitzner:
Thanks! We managed to start looking into this at the same point; I saw the notification when I already had published the new x-revision. https://hackage.haskell.org/package/brittany-0.12.1.1/revisions/

I have included this as f80e777163acfe304ffdfcc209c7872c23a7acfd. I have also removed the dependency entirely because the code was not even using it.. whoops.

PR #288, opened 2020-03-14
Permit extra-1.7

ndmitchell:
I confirmed it compiled and appears to work.


lspitzner:
Thanks, I rebased onto release branch and published as `0.12.1.1.x2`. Github is not clever enough to see it, but this is merged into master.

PR #303, opened 2020-05-05
Readme: Supports 8.8

andys8:
Is it safe to assume it also supports `8.8`? There is a stack file and it seems to work.


lspitzner:
/me at changelog.. support since september '19

Oh, right. Thanks!

PR #311, opened 2020-07-21
Allows aeson-1.5.*

jkachmar:
Verified that this seems to compile and pass tests locally.


georgefst:
> do you mind releasing a new version?

@lspitzner Is there anything holding this up?

Even just a Hackage revision bumping the bound would be very helpful for `haskell-language-server`.

PR #313, opened 2020-08-28
nondecreasing export list formatting

expipiplus1:
```haskell
module Foo
  ( bar
  , baz
  )
where
```

now becomes

```haskell
module Foo
  ( bar
  , baz
  ) where
```

Having the `where` aligned with `module` is a bit weirdy IMO.

Not sure if my patch is the best way to do this, I wrote it a while ago
and have forgotten how much care I put into it.

~Haven't updated the tests yet either, but is this something which would
be merged?~

### DIFF
--- a/src-literatetests/10-tests.blt
+++ b/src-literatetests/10-tests.blt
@@ -1001,8 +1001,7 @@ module Main
   , test7
   , test8
   , test9
-  )
-where
+  ) where
 
 #test exports-with-comments
 module Main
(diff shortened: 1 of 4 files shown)


tfausak:
Looks good to me! I've always wondered by Brittany formats exports lists like this in the first place.

expipiplus1:
Is it possible to get this merged, please?

tfausak:
Sure, I'll go ahead and merge this.

expipiplus1:
Thanks!

PR #323, opened 2020-11-15
Set up GitHub actions

tfausak:


tfausak:
I know that Travis CI is already set up, but GitHub Actions has a couple benefits:

- It's possible to test against Ubuntu, macOS, and Windows. This is also possible on Travis CI, but I feel like the GitHub Actions build runners are quicker and more reliable. 
- Build artifacts, like executables and package tarballs, are first class. 
- Subjectively, it seems like Travis CI is losing mind share and GitHub Actions is gaining it. 

For the time being I'm going to leave Travis CI in place, but it does feel wasteful to have two CI pipelines.

PR #312, opened 2020-08-06
Add support for GHC 8.10.1 

bubba:
Continuing the work of @jneira in #305 , rebased

Issue #48, opened 2017-08-15
Type application with Nats or Symbols upsets let alignment


expipiplus1:
Compare:

```haskell
foo =
  let a = b @1
      cccc = ()
  in  foo

foo =
  let a = b @""
      cccc = ()
  in  foo

foo =
  let a    = b @d
      cccc = ()
  in  foo

foo =
  let a    = b @D
      cccc = ()
  in  foo
```

lspitzner:
Right, this most likely comes down to [this case lacking a proper implementation](https://github.com/lspitzner/brittany/blob/ebf84a03e939ad1017cc741339d0ec7e4080e776/src/Language/Haskell/Brittany/Internal/Layouters/Type.hs#L602-L603). Although I am a bit confused why it does not work as-is; how `layoutByExact` works but inhibits alignment.. will investigate later.

lspitzner:
Fixed on `dev` branch

expipiplus1:
Thanks!

PR #324, opened 2020-11-15
Support GHC 8.10

tfausak:
Continuing work from #312.


tfausak:
Now there is a legitimate test failure against GHC 8.10. It's related to #48. I don't know enough to fix it. 

```
  src-literatetests/Main.hs:183:3: 
  1) regression issue 48 a
       expected: Right 
                 foo =
                   let a    = b@1
                       cccc = ()
                   in  foo
                 
        but got: Right 
                 foo =
                   let a    = _
                       cccc = ()
                   in  foo
                 

  To rerun use: --match "/regression/issue 48 a/"

  src-literatetests/Main.hs:183:3: 
  2) context free: regression issue 48 a
       expected: Right 
                 foo =
                   let
                     a = b@1
                     cccc = ()
                   in foo
                 
        but got: Right 
                 foo =
                   let
                     a = _
                     cccc = ()
                   in foo
```

https://github.com/lspitzner/brittany/pull/324/checks?check_run_id=1403080234#step:7:1202

expipiplus1:
> Now there is a legitimate test failure against GHC 8.10. It's related to #48. I don't know enough to fix it.

Looking at `--dump-full-ast` from brittany with 8.8 and brittany with 8.10, we can see a difference:

where 8.8 returns:

```haskell
[ A Just (Ann (DP (0,-1)) [] [] [] Nothing Nothing)
    GRHS
      NoExt
      []
      A Just (Ann (DP (0,1)) [] [] [((G AnnAt),DP (0,0))] Nothing Nothing)
        EAsPat
          NoExt
          A Just (Ann (DP (0,0)) [] [] [((G AnnVal),DP (0,0))] Nothing Nothing)
            Unqual {OccName: b}
          A Just (Ann (DP (0,0)) [] [] [((G AnnVal),DP (0,0))] Nothing Nothing)
            HsOverLit
              NoExt
              OverLit
                NoExt
                HsIntegral
                  IL
                    SourceText "1"
                    False
                    1
                HsLit
                  NoExt
                  HsString
                    SourceText "noExpr"
                    {FastString: "noExpr"}
]
```

8.10 returns a hole!

```haskell
[ A Just (Ann (DP (0,-1)) [] [] [] Nothing Nothing)
    GRHS
      NoExtField
      []
      A Just (Ann (DP (0,1)) [] [] [((G AnnVal),DP (0,0))] Nothing Nothing)
        HsUnboundVar
          NoExtField
          TrueExprHole {OccName: _}
]
```

It's also not valid syntax as ghci tells us:

```
λ> b = a@()

<interactive>:3:5: error:
    Pattern syntax in expression context: a@()
    Type application syntax requires a space before '@'
```

I think it's fine to remove this test (or correct it so that it has the space before the `@`)

expipiplus1:
It's surprising to me that the AST can even represent this, although perhaps it's useful for error messages or something

tfausak:
> or correct it so that it has the space before the `@`

👍

expipiplus1:
Need to update the supported GHC versions in the readme

tfausak:
I should make sure there are `stack.yaml` files for each configuration. Then I think this will be ready to go.

expipiplus1:
@tfausak would you like any help with this release?

tfausak:
Sorry! Thanks for the ping. I don't need any help with this, I just have to actually do it. With any luck I should be able to spend some time on it in the next few days.

PR #326, opened 2020-11-20
Simplify Nix expressions

expipiplus1:
This uses `developPackage` from `nixpkgs` to generate the derivation

Also bump the pinned nixpkgs version

Also add CI for nix build

Opened against the ghc 8.10 branch because it includes those changes.

PR #329, opened 2020-12-10
Simplify Nix expressions

expipiplus1:
This uses `developPackage` from `nixpkgs` to generate the derivation

Also add CI for nix build


expipiplus1:
reopened from https://github.com/lspitzner/brittany/pull/326#event-4092913743

default.nix
```diff
@@ -1,38 +1,18 @@
...
-  license = stdenv.lib.licenses.agpl3;
+
+let
+  haskellPackages = if compiler == null then
+    pkgs.haskellPackages
+  else
+    pkgs.haskell.packages.${compiler};
+
+in haskellPackages.developPackage {
+  name = "";
```
tfausak, line 15: Should this have `name = "brittany"`, or does it not matter?

tfausak (approved):
I am not a Nix wizard but this does indeed look like an improvement. Thanks!

default.nix
expipiplus1: It doesn't *really* matter, in this case it'll use the enclosing directory name. However it should be "brittany" to make sure the same derivation is generated regardless of the directory name. Fixed.

expipiplus1:
Thanks!

PR #334, opened 2021-01-02
Update ArchLinux installation instructions

ahstro:
Brittany is in the community repo now, and no longer requires AUR


tfausak (approved):
Thanks!

PR #363, opened 2022-01-07
Fix warnings when building with GHC 9.0.2

sergv:


tfausak:
Oof, that's unfortunate. I suppose that's what I get for manually initializing `DynFlags`. 

Do you think it would be possible to use [defaultDynFlags](https://hackage.haskell.org/package/ghc-9.0.2/docs/GHC-Driver-Session.html#v:defaultDynFlags) instead?

sergv:
I'll give `defalutDynFlags` a try.

sergv:
@tfausak I have switched the module to using `defaultDynFlags`, now there's no need for preprocessor. It seems to work - all the tests pass.

tfausak (approved):
Awesome, thank you!

